PR #2, opened 2019-07-12
Day 8 - Problem 1 and 2 done

Hariae:
Problem 2 : Stuck in forming the DP formula.

### DIFF
--- a/CountCoinChange.java
+++ b/CountCoinChange.java
@@ -1,0 +1,45 @@
+class CountCoinChange {
+    public int change(int amount, int[] coins) {
+        int[][] dp = new int[coins.length+1][amount+1];
+
+        //fill 1st column as 0
+        for(int i=0;i<dp.length;i++){
+            dp[i][0] = 0;
+        }         
+
+        //fill first row as max
+        for(int j=1;j<dp[0].length;j++)
+            dp[0][j] = 9999;
+
+        //calc dp array
+        for(int i=1;i<dp.length;i++){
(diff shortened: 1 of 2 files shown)


CountCoinChange.java
```diff
@@ -12,21 +12,23 @@ public int change(int amount, int[] coins) {
...
-            System.out.println();
+        // for(int i=1;i<dp.length;i++){
+        //     for(int j=1;j<dp[0].length;j++){
+
+        //         //copy from top
+        //         if(j < coins[i-1]){
+        //             dp[i][j] = dp[i-1][j];
+        //         }
+        //         //compare and calc
+        //         else{
```
ashmeetdeol19: You are doing something different here . Right ?

PaintHouses.java
```diff
@@ -0,0 +1,36 @@
...
+            {3,6,1},
+            {8,7,4},
+            {6,2,9},
+            {1,4,7},
+            {2,3,6},
+        };
+
+        PaintHouses obj = new PaintHouses();
+        System.out.println(obj.paintHouses(costs));
+    }
```
ashmeetdeol19: Thats a Great Solution 👍

CountCoinChange.java
Hariae: Yes, Ashmeet. I started out thinking I should do some steps after filling the matrix (like a traversal to determine number of combinations). Then realized the formula to fill dp matrix is different, so updated it in the final commit.

PR #4, opened 2019-07-12
Please review Day8 problems

MahiteeS:


CountChange2.java
```diff
@@ -0,0 +1,46 @@
...
+
+//Your code here along with comments explaining your approach
+public class CountChange2 {
+	static long countWays(int S[], int m, int n) 
+    { 
+        // table[i] will be storing the number of solutions 
+        // for value i. We need n+1 rows as the table is 
+        // constructed in bottom up manner using the base 
+        // case (n = 0) 
+        long[] table = new long[n+1];
```
ashmeetdeol19: Why you made this as long ?
If you want to return int array, then you will need a new array where you can copy all values after type casting.

HousePaint.java
```diff
@@ -0,0 +1,34 @@
+package day8;
+
+//Time Complexity : O(n) where n is the length of house array
+//Space Complexity : O(1) 
+//Did this code successfully run on Leetcode : can't try because the question was locked
+//Any problem you faced while coding this : This solution is understandable but it is returning hash value
```
ashmeetdeol19: Can you explain this ?

PR #7, opened 2019-07-12
Updated Problem 1 and 2, Day 8

gauravj09:


problem2.cpp
```diff
@@ -0,0 +1,36 @@
...
+            for(int j =1; j<amount+1 ; j++)
+            {
+                if(j<coins[i-1])
+                    dp[i][j] =  dp[i-1][j] + 0; 
+                else  
+                    dp[i][j] = dp[i-1][j] + dp[i][j - coins[i-1]];
+            }
+
+        return dp[coins.length][amount];
+    }
```
ashmeetdeol19: Well Done Gaurav 👍

problem2.cpp
```diff
@@ -15,15 +15,15 @@ using namespace std;
...
+        int dp[coins.size()+1][amount+1];
 
         dp[0][0] = 1;
 
-        for(int i =1; i<coins.length+1; i++)
+        for(int i =1; i<coins.size()+1; i++)
             dp[i][0] = 1;
 
-        for(int i =1; i<coins.length+1;i++)
+        for(int i =1; i<coins.size()+1;i++)
```
ashmeetdeol19: Why did you replace .length by .size() ?

gauravj09: I was looking at a Java code for reference and I forgot to change the .length syntax of Java to .size() syntax of c++ while pushing the previous commit.

PR #22, opened 2019-08-31
Problem 1: paint n row 3 column houses

manishjangid:


manishjangid:
@kaurramanpreet  Yes, I have already written it in problem description in the program. From next time I will also mention it in pull request.
Thanks for the feedback.

PR #25, opened 2019-09-01
Paint house and Coin change 2 done

pradhan1234:


pradhan1234:
cc @kaurramanpreet added falling path sum, please review :)

PR #33, opened 2019-10-18
Completed -Please Review

sindhu819:


Python-21.py
```diff
@@ -0,0 +1,26 @@
...
+        c=len(costs[0])
+        dp=[[0]*c for _ in range(r)]
+        dp[0][0]=costs[0][0]
+        dp[0][1]=costs[0][1]
+        dp[0][2]=costs[0][2]
+        
+        #edge case
+        if not costs:
+            return 0
+
```
sehaj-grover: try checking the egde cases in the start

sindhu819: Changed the edge case. Thank you!

Python-21.py
```diff
@@ -0,0 +1,26 @@
...
+        # write your code here
+        r=len(costs)
+        c=len(costs[0])
+        dp=[[0]*c for _ in range(r)]
+        dp[0][0]=costs[0][0]
+        dp[0][1]=costs[0][1]
+        dp[0][2]=costs[0][2]
+        
+        #edge case
+        if (costs == [] or len(costs) == 0):
```
sehaj-grover: what I meant was that the edge cases should be checked at the top. If costs was null you will receive exception in the start where accessing cost because the check for costs comes later

sindhu819: ok got it.

PR #40, opened 2019-11-28
Adding DP2 changes

dasgprat:
Problem 2 not returning correct solution
Will submit after class

### DIFF
--- a/Problem1_Optimal.java
+++ b/Problem1_Optimal.java
@@ -1,0 +1,70 @@
+ 
+/**
+LeetCode Submitted : NA (Not from Leetcode)
+Space Complexity : O(1)
+Time Complexity : O(N)
+ 
+At each level of decision tree three choices need to be made either paint house RED,GREEN or BLUE. Thus, accordingly select next set of houses with different colors. Repeat this for all the houses. Now instead of recursively calling each overlapping sub-problems we store these values in a 3 variables storing max cost if painted house with that color
+**/
+ 
+class Main {
+ public static void main(String[] args) {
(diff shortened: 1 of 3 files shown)


Problem1_Optimal.java
```diff
@@ -0,0 +1,70 @@
...
+     //cost_matrix[i][2] = hcost[i][2] + Math.max(cost_matrix[i-1][0],cost_matrix[i-1][1]); //choosing BLUE color
+    
+     blue = hcost[i][2] + Math.max(temp,temp1);
+ 
+ }
+  //int max =  Math.max(cost_matrix[size-1][0],Math.max(cost_matrix[size-1][1],cost_matrix[size-1][2]));
+ 
+ int max = Math.max(red,Math.max(green,blue));
+ System.out.println("Maximum cost of Painting the houses = " + max);
+
```
neha79091: You have to find minimun cost. You are using max here.

Problem2_Optimal.java
```diff
@@ -0,0 +1,56 @@
+
+/**
+LeetCode Submission :
+Space Complexity : O(N^2)
+Time Complexity : O(N^2)
+Use similar logic from previous problem once we enounter a solution i.e. amount = 0 add 1 to final possible solution
```
neha79091: Time and space complexities are incorrect.

dasgprat: oh  I thought it will be O(Amount * Number of Denominations available) ?

neha79091: yes that's correct.

PR #41, opened 2019-11-29
Added solution for CoinChange2 and HousePaint

amitkc1:


CoinChange2.java
```diff
@@ -0,0 +1,28 @@
+// Time Complexity : o(S*n), where S is the amount
+// Space Complexity :O(S), where s is the amount
+// Did this code successfully run on Leetcode : Yes
+//Approach: Problem is solved using dynamic programming approach.
```
neha79091: Space complexity will be O(S*n) because you are making a matrix here.

PR #66, opened 2020-01-08
submitting DP-2

samhithaaaa:


painhouse.py
```diff
@@ -0,0 +1,22 @@
+# time complexity = O(n)
+# space complexity = O(1)
+class Solution(object):
+    def minCost(self,costs):
+        #baseCondition
+        if not costs:
+            return None
```
neha79091: You have to return 0 here

PR #78, opened 2020-02-16
after class solutions - please review

pbandyop:
@Lohith-Aswathappa


Lohith-Aswathappa:
Great work Payel!

Please add inline comments, it'll be easier to go through later

PR #100, opened 2020-03-26
solved problem 1

schakrad:


DP-2-PROBLEM-1.txt
```diff
@@ -0,0 +1,21 @@
+#Problem1
+#Time Complexity : o(3*N), in general o(M*N) for M colors
```
devashishnyati: Please explain the time complexity? Why is it O(m*m)?

devashishnyati:
Good job Sindhu!
Submit other problems also.

devashishnyati:
Also, submit the brute force solution.

PR #103, opened 2020-03-27
Completed DP2 - Please Review

kannan-mrugapphan:
1. The Coin Change 2 problem passes all test cases on leetcode.
2. The Paint House problem was tested only against a few cases since it was a premium problem.


devashishnyati:
Good job Kannan!
Please submit the Brute Force solution also.

PR #104, opened 2020-03-28
Completed DP-2 initial version - Please review

jose-rafael-marcano:


devashishnyati:
Good job Jose!
Please submit the Brute Force solution also.

PR #108, opened 2020-04-30
DP solution for both problems

domzhaomathematics:


sowmyadvn:
Both paint house and coin change solutions/complexities are correct. Good job Dominic!

sowmyadvn:
Both paint house and coin change solutions/complexities are correct. Good job Dominic!

PR #109, opened 2020-05-01
Completed brute force approaches

namrataKasar:


Sample.java
```diff
@@ -1,7 +1,71 @@
...
+            }
+            else {
+                price += Math.min(helper(costs, i+1, 0), helper(costs, i+1, 1));
+            }
+        }
+        
+        return price;
+    }
+
+    // Time Complexity : O(2^n)
```
sowmyadvn, line 39: Please check time complexity here. We have two factors that affect it, coin denominations and amount

sowmyadvn:
Good job with brute force solutions Namrata! Try solving them with DP as well. Please check in-line comments

sowmyadvn:
Good job with brute force solutions Namrata! Try solving them with DP as well. Please check in-line comments

PR #114, opened 2020-05-02
Completed Design for Coin change question

JacobStephen9999:


iivaishnavii:
Good job, Jacob!Submit the other problem as well.

PR #121, opened 2020-05-07
Update Sample.java

ashutosh-senapati:


super30admin:
Brute force solutions missing for both problems

PR #122, opened 2020-05-07
DP-2

adityad30:


21PaintHouse.py
```diff
@@ -0,0 +1,35 @@
+"""
+// Time Complexity :O(mn) where m->cost and n->colours
```
sowmyadvn, line 2: Because we have only 3 colors, O(3m) would be the complexity, which is ~O(m)

21PaintHouse.py
```diff
@@ -0,0 +1,35 @@
...
+// Space Complexity :O(n)
```
sowmyadvn, line 3: If n is colors based on your comment above, O(m) would be space complexity

22CoinChange.py
```diff
@@ -0,0 +1,31 @@
+"""
+// Time Complexity :O(n2) where n is denominations and money.
```
sowmyadvn, line 2: Because amount(m) and coins(n) are different variables, complexity would be O(mn)

22CoinChange.py
```diff
@@ -0,0 +1,31 @@
...
+// Space Complexity :O(1)
```
sowmyadvn, line 3: We are returning only int, but utilized an array. So, do you think auxiliary space here is still constant?

sowmyadvn:
Please review the inline comments Aditya. We could discuss further on time and space complexities of the above problems. Feel free to ping me on slack

sowmyadvn:
Please review the inline comments Aditya. We could discuss further on time and space complexities of the above problems. Feel free to ping me on slack

22CoinChange.py
adityad30: I missed few comments, sorry about that. I will recheck all the DP code and other submitted PR's and will update the time and space complexity accordingly. Thank you.

PR #124, opened 2020-05-07
Completed DP-2 - Please review

khushal1198:


DP_2_Problem_1.java
```diff
@@ -0,0 +1,73 @@
+//Brute force
+//Time Complexity : O(2^n * 3))
+//Space Complexity : Just the stack space O(1)
+// Have not bought leetcode premium yet
+// Problems :  No problem
+
+//Optimized DP
+//Time Complexity : O(n)
+//Space Complexity: O(m*n)
```
sowmyadvn, line 9: Because colors are constant i.e. 3, there is only one factor affecting space, which is number of houses. So, space would also be O(N)

sowmyadvn:
Solutions are correct Khushal! Please check inline comment

sowmyadvn:
Solutions are correct Khushal! Please check inline comment

PR #128, opened 2020-05-08
Completed DP-2

shalinych:


super30admin:
Brute force solutions are missing for both. DP solutions are correct

PR #135, opened 2020-06-04
Please comment about the complexities for the recursive solution.

trivedipratul:


PA_21.py
```diff
@@ -0,0 +1,27 @@
+#LC 256 - Paint House
+#Time Complexity - O(n) Not sure, can you comment?
```
Lohith-Aswathappa, line 2: Since you are using recursion, the time complexity will be exponential
In this case, cost power color

PA_22.py
```diff
@@ -0,0 +1,66 @@
+#LC 518 - Coin Change 2
+#Recursive Solution
+#Time Complexity - N^m where N is the number of coins and m is the amount
```
Lohith-Aswathappa, line 3: Again, 2 power n Since you are making 2 choices at each point

PA_22.py
```diff
@@ -0,0 +1,66 @@
+#LC 518 - Coin Change 2
+#Recursive Solution
+#Time Complexity - N^m where N is the number of coins and m is the amount
+#Space Complexity - O(n) where n is the amount
```
Lohith-Aswathappa, line 4: This seems to be correct

PA_21.py
Lohith-Aswathappa: Try the iterative solution for this as well.

PR #151, opened 2020-06-05

akshatasawhney:


Problem1.py
```diff
@@ -0,0 +1,68 @@
+"""
+// Time Complexity : o(m*3)
```
Lohith-Aswathappa, line 2: is it m power 3?

Problem1.py
```diff
@@ -0,0 +1,68 @@
...
+// Space Complexity : o(1)
```
Lohith-Aswathappa, line 3: Add space complexity considering recursive stack

Problem2.py
```diff
@@ -0,0 +1,63 @@
+"""
+// Time Complexity : o(m*n)
+// Space Complexity : o(m*n)
+// Did this code successfully run on Leetcode : yes
+// Any problem you faced while coding this : no
+
```
Lohith-Aswathappa, line 6: add time and space complexities for brute force as well.

Lohith-Aswathappa:
Good work, Akshata!
review the inline comments.

Problem1.py
akshatasawhney: No, its m multiplied by 3, where m is the number of houses and 3 for the colors.

PR #152, opened 2020-06-05
Paint hourse DP solution

mithi21:


AroopN:
Good work on uploading both bruteforce and DP solutions. 

Please verify the time and space complexity of coin change brute force solution.
Also, in Paint House DP solution, the space complexity is O(1) as we are modifying the input itself.

AroopN:
Good work on uploading both bruteforce and DP solutions.

Please verify the time and space complexity of coin change brute force solution.
Also, in Paint House DP solution, the space complexity is O(1) as we are modifying the input itself.

PR #157, opened 2020-06-05
Create PaintHouse.py

Shashank001122:


Lohith-Aswathappa:
Good job, Shashank!
Push the other problem and add time and space complexities for both.

PR #171, opened 2020-06-10
DP-2--Completed

VirajDeshmukh1:


paint-house.java
```diff
@@ -0,0 +1,24 @@
+// Time Complexity : O(n*m)
+// Space Complexity : O(n*m)
```
AbhiniveshP, line 2: The modifications to the cost matrix are in-place and so the Space complexity would be O(1)

AbhiniveshP (commented):
Nice work on both the questions. Please check the space complexity for min_cost question.

AbhiniveshP:
Nice work on both the questions. Please check the space complexity for min_cost question.

PR #195, opened 2020-07-19
completed, please review

JaneJianLee:


solution1.py
```diff
@@ -0,0 +1,28 @@
+# https://leetcode.com/problems/paint-house
+
+# // Time Complexity : o(n*3) n= number of houses
+# // Space Complexity : o(n*3) n= number of houses
```
samarthjj, line 4: The time and space should only be O(n), the 3 is a constant and can be ignored

samarthjj:
Good work Jane, Please review the inline comments!
Please also complete the **Recursive/Greedy** solution as well!
Also please add your 3-4 bullet points of approach for every question!

samarthjj:
Good work Jane, Please review the inline comments!
Please also complete the **Recursive/Greedy** solution as well!
Also please add your 3-4 bullet points of approach for every question!

PR #241, opened 2020-08-31
Problem 1 & 2 - Completed - Please review

desham05:


khushal1198 (commented):
Nice job with approach and code Sham!

khushal1198:
Nice job with approach and code Sham!

PR #268, opened 2020-09-25
Paint house problem done

meetdave06:


paintHouse.py
```diff
@@ -0,0 +1,51 @@
+# Time complexity :- O(n)
```
kannan-mrugapphan, line 1: Shouldn't the time be exponential here

kannan-mrugapphan (commented):
Good Work. Please submit the other problem as well

kannan-mrugapphan:
Good Work. Please submit the other problem as well

PR #323, opened 2020-12-09
DP-2 Complete. Please Review. Thank you

Vasundhara-Krishna:


house_paint.cpp
```diff
@@ -0,0 +1,20 @@
+// Time Complexity : O(n)
+// Space Complexity : O(1)
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this : Yes
```
super30admin, line 4: What was the problem faced. Please mention/dm

super30admin (commented):
Please submit the other solutions as well for both problems. These optimized versions are correct

super30admin:
Please submit the other solutions as well for both problems. These optimized versions are correct

PR #349, opened 2021-01-05
dp-2 - Please Review

gangakalburgi:


CoinChange2.java
```diff
@@ -0,0 +1,56 @@
+// Time Complexity : O(n^2), n is the number of coins and amount
```
Lohith-Aswathappa, line 1: n cannot be the same for both coins and amount 
Your complexity should be O(MxN)

PaintHouse.java
```diff
@@ -0,0 +1,79 @@
...
+        
+          return Math.min(helper(costs,row+1,0,minAmount+costs[row][2]),
+                          helper(costs,row+1,1,minAmount+costs[row][2]));
+       
+        
+    } */
+        
+        //Method 2:
+        //dp approach 
+        //time complexit: O(n^2)
```
Lohith-Aswathappa, line 41: columns is fixed in this, Complexities should be the number of rows

Lohith-Aswathappa (commented):
Good work, Ganga!
Check inline comments

Lohith-Aswathappa:
Good work, Ganga!
Check inline comments

PR #357, opened 2021-01-06
submitting dp-2 assignment

yahska-ednihs:


paint_house.py
```diff
@@ -0,0 +1,29 @@
+# Time Complexity : O(m*n) where m and n are dimensions of costs matrix
+# Space Complexity : O(m*n)
```
samarthjj, line 2: Time & Space is O(n), because the nested arrays are always size 3!

samarthjj:
Good work Akshay, Please review the inline comments.
Please also submit the recursive solution for all DP problems too.

samarthjj:
Good work Akshay, Please review the inline comments.
Please also submit the recursive solution for all DP problems too.

PR #383, opened 2021-01-29
Completed DP-2 : Please Review

Amrindersingh1:


minCost.java
```diff
@@ -0,0 +1,34 @@
+// Time Complexity : O(m*n)
```
nsibal, line 1: It should just be O(n) (n = number of houses).

nsibal (commented):
Look at the inline comment of `minCost.java`.

Otherwise, looks fine!

nsibal:
Look at the inline comment of `minCost.java`.

Otherwise, looks fine!

PR #392, opened 2021-02-01
Add files via upload

EnnaSachdeva:


coin_change_2.py
```diff
@@ -0,0 +1,31 @@
+#Tume complexity: O(n*T): where n: # of coins, T: target amount value
+#Space complexity: O(n*T)
+#This algorithm uses dynamic programming, by converting it into several subproblems, where each subproblem solves for
+# finding the number of ways to reach a target values, by first picking certain number of coins.
+# We keep on adding more number of coins, till we cover all coins, and target value is reached.
+
+import numpy as np
```
nsibal, line 7: I am not sure, but I don't think you'd be allowed to use external libraries. Please verify this with your instructor.

paint_house.py
```diff
@@ -0,0 +1,19 @@
+#Tume complexity: O(n): where n: # of houses
+#Space complexity: O(n)
```
nsibal, line 2: Shouldn't your SC be O(1)? You haven't used any extra space. You just solved this question in-place, which in my opinion, you shouldn't have.

nsibal (commented):
Read inline comments, otherwise looks alright!

nsibal:
Read inline comments, otherwise looks alright!

PR #411, opened 2021-02-28
Completed Dp-2

svyas97:
Please check the TC and SC, I think it might be wrong

### DIFF
--- a/choinchange.py
+++ b/choinchange.py
@@ -1,0 +1,38 @@
+        for k in range(m):
+
+                                if (k != j):
+                                                            MIN =   min (nums[i-1][k], MIN) #Comparing for min with all the values except the value
+                                                                                    #in the same columns
+
+
+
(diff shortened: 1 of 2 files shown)


nsibal (commented):
Please fix the indentation of the code. Also, just by looking at the first few lines, I think your code won't execute because `j` isn't defined before L3 in both the files.

nsibal:
Please fix the indentation of the code. Also, just by looking at the first few lines, I think your code won't execute because `j` isn't defined before L3 in both the files.

svyas97:
> Please fix the indentation of the code. Also, just by looking at the first few lines, I think your code won't execute because `j` isn't defined before L3 in both the files.

Hi Nirbhay, looks like the file has some issues, I will create a new pull request.

svyas97:
Created a new pull request

PR #471, opened 2021-06-04
Completed problem 2 and 1

kgajwan1:
Don't have more test cases for problem 1, so made as per interpretation


keshav-Gupta-IS (commented):
Good Work, Karan!! Just correct the paint house problem it will fail  for some of the cases.

keshav-Gupta-IS:
Good Work, Karan!! Just correct the paint house problem it will fail  for some of the cases.

PR #492, opened 2021-06-07
Completed

s4git21:


ramya-sruthi-d (commented):
Good job with the code, the tc is O(n^2) instead of O(2^n), both of them are not the same and mean a lot different.

ramya-sruthi-d:
Good job with the code, the tc is O(n^2) instead of O(2^n), both of them are not the same and mean a lot different.

ramya-sruthi-d:
Ignore the comment, I thought the tc is for the optimal solution, sorry for the confusion.

PR #557, opened 2021-07-05
optimum code completed.

venkatagudimela:
completed with optimal code please review


Rooppesh:
Good Job with the code. Please add TC and SC along with the comments.

Rooppesh:
Good Job with the code. Please add TC and SC along with the comments.

PR #697, opened 2021-12-30
Completed DP-2. Please review

Divicodes:


sonalgupta276 (commented):
Good job! Can also do recursive approach first

sonalgupta276:
Good job! Can also do recursive approach first

PR #899, opened 2022-06-06
DP-2

nishanth-sayana:
I am having doubt in the Space Complexity of CoinChange-2. Please let me know about the Space Complexity of that problem


niharikagupta2021 (commented):
Good Job! In coin change-2 you are using a 2d array of size m*n, so sc will be O(m*n)

niharikagupta2021:
Good Job! In coin change-2 you are using a 2d array of size m*n, so sc will be O(m*n)

PR #920, opened 2022-07-02
Completed DP-2.

viveksharma180:
Signed-off-by: viveksharma180 <sharma.viv@northeastern.edu>

Please review DP-2.

### DIFF
--- a/PaintHouse.java
+++ b/PaintHouse.java
@@ -1,0 +1,31 @@
+// Time Complexity : O(N) or O(n*3) where n are houses and 3 is colors
+// Space Complexity : O(N)
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this : No
+
+
+// Your code here along with comments explaining your approach
+
+
+public int minCost(int[][] costs) {
+    int m = costs.length;
+    int n=3;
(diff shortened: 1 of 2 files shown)


PaintHouse.java
```diff
@@ -0,0 +1,31 @@
+// Time Complexity : O(N) or O(n*3) where n are houses and 3 is colors
```
jeelpatel18, line 1: It will be O(N^2)

jeelpatel18 (commented):
GOOD JOB

jeelpatel18:
GOOD JOB

PaintHouse.java
viveksharma180: Hi Jeel,
In the second loop, n is constant i.e. 3. So shouldn't it be O((m-1)*3) or O(N)?

jeelpatel18: Yes, nested loop is there right!!! but although best case will be 3*m-1 as you  said.

jeelpatel18: After expanding the term it is O(N) .

viveksharma180: Understood. Thanks for the clarification Jeel.

PR #921, opened 2022-07-02
Both solutions

pranavsinghps1:


house_paint.py
```diff
@@ -0,0 +1,13 @@
+"""
+TC:O(n)
+SC:O(n)
```
jeelpatel18, line 3: You ar not using any extra space so SC is O(1) here.

jeelpatel18 (commented):
Well done.

jeelpatel18:
Well done.

PR #929, opened 2022-07-03
completed

shrutinag04:


housePaint.java
```diff
@@ -0,0 +1,28 @@
+/**Time Complexity :
+    O(n)
+// Space Complexity :
+    we are using same matrix hence O(n)
```
ishika2741, line 4: Since we are using same matrix , so there is no auxiliary space used . So space would be O(1)

ishika2741 (commented):
Well done!

ishika2741:
Well done!

PR #931, opened 2022-07-04
DP-2

chaitanya3111998:


Problem1.cpp
```diff
@@ -0,0 +1,25 @@
+// Problem1(https://leetcode.com/problems/paint-house/)
+// Time Complexity : O(m) where m is the size of the matrix
```
daredevilteja, line 2: here m is number of rows

daredevilteja (commented):
Good Work! Keep it up

daredevilteja:
Good Work! Keep it up

PR #942, opened 2022-07-06
Completed Dp-2 _Please Review

rajdshah:


PaintHouse2.py
```diff
@@ -0,0 +1,18 @@
+#Time Complexity : O(MN)
```
jeelpatel18, line 1: O(N) will be the TC

jeelpatel18 (commented):
Good job Raj

jeelpatel18:
Good job Raj

PR #952, opened 2022-07-28
dp-2

madhuri434:


coin_change2.java
```diff
@@ -0,0 +1,58 @@
+
+class Solution{
+    public int coinChange(int amount, int[] coins){
+        if(coins == null || coins.length == 0) return 0;
+        //row of the DP matrix
+        int n = amount;
+        int m = coins.length;//error: cannot find symbol [in __Driver__.java]
```
Sumesh-H, line 7: Look at the method name, if you're running this on LeetCode, the main method won't find it (Compiler error). Correct it to public int change (int amount, int[] coins) {}
Solution is correct!

Sumesh-H (commented):
Good Job! Please address the inline comment.

Sumesh-H:
Good Job! Please address the inline comment.

PR #1004, opened 2022-09-14
Completed both problems for DP-2

ysoni0303:


pvk3081 (commented):
Good job. kindly include brute force for paint house problem and check time complexity for 2nd problem

pvk3081:
Good job. kindly include brute force for paint house problem and check time complexity for 2nd problem

PR #1015, opened 2022-09-21
Complete

deepmahajan987:


problem2.java
```diff
@@ -0,0 +1,32 @@
+public class problem2 {
+
+
+// Time Complexity : O(n)
```
daredevilteja, line 4: TC and SC is wrong

daredevilteja:
Good Work! Keep it up

PR #1068, opened 2022-11-08
Solved DP 2 both questions. Please review it.

dharatipatel123:


PaintHouse.java
```diff
@@ -0,0 +1,73 @@
+package S30_Codes.DP_2;
+import java.util.Arrays;
+
+// Time Complexity : O(house*color*color)
```
daredevilteja, line 4: recheck TC and SC

daredevilteja:
Good Work! Keep it up

PR #1189, opened 2023-06-05

ghost:


problem1.cpp
```diff
@@ -1,4 +1,26 @@
 # Time Complexity : O(cn)
```
ishika2741, line 1: Kindly check TC. Also tell what c and n means.

problem2.cpp
```diff
@@ -0,0 +1,20 @@
+# Time Complexity : O(c.n)
```
ishika2741, line 1: Kindly explain what c and n means.

ishika2741 (commented):
Good work! Try to submit exhaustive approaches as well.

ishika2741:
Good work! Try to submit exhaustive approaches as well.

PR #1216, opened 2023-07-10
Completed DP-2 - Please review

Hemanth-Reddy98:


Problem2.py
```diff
@@ -0,0 +1,14 @@
+class Solution:
+    # Time complexity: O(amount * len(coins))
+    # Space complexity: O(amount)
```
ishika2741, line 3: Since you are using a 2-D array here , so it will be amount*len(coins)

ishika2741 (commented):
Well done! Keep it up.

ishika2741:
Well done! Keep it up.

PR #1274, opened 2023-10-07
DP-2 Complete

harshpatel226:


Nandha2Tech (commented):
Great Work! For paint house SC would be O(m*n)

Nandha2Tech:
Great Work! For paint house SC would be O(m*n)

PR #1278, opened 2023-10-08
Done DP 2

ajithvarma97:


Nandha2Tech (commented):
Great Work!

Nandha2Tech (commented):
Great Work! 
In paint house problem, kindly check the inside the first if condition 
It should be matrix[i][j] += Math.min(matrix[i-1][j+1],matrix[i-1][**j+2**]); 
For coin change 2 problem TC and SC would be O(m*n).

Nandha2Tech:
Great Work! 
In paint house problem, kindly check the inside the first if condition 
It should be matrix[i][j] += Math.min(matrix[i-1][j+1],matrix[i-1][**j+2**]); 
For coin change 2 problem TC and SC would be O(m*n).

PR #1399, opened 2024-06-01
completed DP-2

gsprasannabalaji:


Nandha2Tech (commented):
Great Work! For paint house recursive logic TC would be exponential O(2^n) where n is the number of houses.

Nandha2Tech:
Great Work! For paint house recursive logic TC would be exponential O(2^n) where n is the number of houses.

PR #1405, opened 2024-06-06
solution to DP-2

yash18082002:


Nandha2Tech (commented):
Great Work! For paint house problem no extra space is used in your code. So SC would be O(1).

Nandha2Tech:
Great Work! For paint house problem no extra space is used in your code. So SC would be O(1).

PR #1450, opened 2024-08-05
solved both

nehayj100:


Nandha2Tech (commented):
Great Work! TC for coin change would be O(m * n).

Nandha2Tech:
Great Work! TC for coin change would be O(m * n).

PR #1478, opened 2024-10-13
solved

SriVinayA:


Erlybird:
Good job! Are your solutions working on Leetcode? I'm receiving a type error for the result because you're using NumPy. I cast the result to int for both solutions, and they worked fine afterward. Please cross-check this.

Erlybird:
Good job! Are your solutions working on Leetcode? I'm receiving a type error for the result because you're using NumPy. I cast the result to int for both solutions, and they worked fine afterward. Please cross-check this.

PR #1538, opened 2025-01-11
solved DP - 2

csk731:


Erlybird:
Good job with the solutions, it would be better if you follow the input parameters of the problem, instead of adding your parameters for your convenience.

Erlybird:
Good job with the solutions, it would be better if you follow the input parameters of the problem, instead of adding your parameters for your convenience.

PR #1601, opened 2025-02-26
DP-2 Completed

vs-1612:


paint-house.py
```diff
@@ -0,0 +1,15 @@
+#Time Complexity = O(n)
+#Space Complexity: O(1)
```
ishika2741, line 2: Dp array would take space

ishika2741 (commented):
Well done!

ishika2741:
Well done!

PR #1688, opened 2025-07-01
DP-2 Done

NUPUR0307:


super30admin:
Strengths:
- Good understanding of DP concepts and ability to optimize from exhaustive approaches
- Clear comments explaining the approach
- Correct implementation for Problem2

Areas for Improvement:
- Be more careful with loop conditions and edge cases (Problem1 has an infinite loop bug)
- Consider using separate variables instead of modifying input arrays for better practice
- For Problem2, the final ternary operator is unnecessary since dp[coins.length][amount] will never be negative

super30admin:
Strengths:
- Good understanding of DP concepts and ability to optimize from exhaustive to DP solutions.
- Clear comments explaining the approach.
- Efficient space usage in Problem1.

Areas for Improvement:
- Careful attention to loop conditions and increments is crucial. The `i++` in Problem1 would cause an infinite loop.
- In Problem2, the ternary operator at the end (`== 0 ? 0 : ...`) is unnecessary since dp[coins.length][amount] will always be the correct count (including 0 if no combinations exist).
- Consider adding edge case handling in Problem1 (e.g., empty costs array).

PR #1714, opened 2025-08-05
completed DP-2

DPoo1:


super30admin:
- The main issue is that the solution does not address the problem described (DP-2). It's important to carefully read and understand the problem statement before starting to code.
- The approach and implementation for the Coin Change 2 problem are correct and well-explained. The use of dynamic programming with a matrix is appropriate for this problem.
- The comments are helpful and explain the thought process clearly, which is good practice.
- For future submissions, ensure that the solution matches the problem description. If there was confusion about the problem, it would be good to clarify before proceeding.

DPoo1:
Thank you for the insight, as always. If you don't mind, I am not understanding. Could you please explain what you meant by " the solution does not address the problem described (DP-2) ". Did you mean that I did not explain why I used DP, or that I did not explain the problem statement? I would greatly appreciate it if you can guide me on how to correctly do the submission per say- where to have the approach/ how to properly structure the approach along with time and space complexities. Thank you for your attention and guidance.

PR #1726, opened 2025-08-29
DP-2 done

desaiparam:


super30admin:
- Strengths: The student has implemented an efficient dynamic programming solution with optimal time and space complexity. The code is clean, well-commented, and easy to understand.
- Areas for Improvement: While the solution is correct, it might be helpful to add a brief explanation of why the iteration is done from the last house to the first (bottom-up approach) in the comments. This would make the reasoning behind the approach clearer to someone reading the code.

PR #1739, opened 2025-09-03
Solve problem 1 and 2

sarvesh-gaurishankar-sawant:


super30admin:
- The Solution1.java is well-implemented and efficient. The dynamic programming approach is correctly used to solve the problem.
- The comments are clear and helpful, but it would be better to mention the problem statement in the comments for clarity.
- The variable names are descriptive, which makes the code easy to understand.
- The Solution2.java is not relevant to the DP-2 problem and should not be included in the submission for this problem.

sarvesh-gaurishankar-sawant:
"The Solution2.java is not relevant to the DP-2 problem and should not be included in the submission for this problem." Solution2.java is for coin change 2  @super30admin

PR #1740, opened 2025-09-10
completed DP-2

aditya18m:


super30admin:
- Problem1.cpp: The solution is excellent. The approach is efficient and the code is clean. The use of temporary variables to store intermediate results is a good practice.
- Problem2.cpp: The solution is also excellent. The use of `unsigned __int128` to handle large numbers is a good approach, though it's worth noting that this might not be portable across all platforms. The comments and code structure are clear and easy to follow.

Overall, the student has done a great job in solving both problems efficiently and correctly. The code quality is high, and the solutions are well-optimized.

PR #1782, opened 2026-01-16
Added Solutions to both DP-2 Problesms

joshiatharv1:


super30admin:
Strengths:
- The solution correctly implements the dynamic programming approach to solve the problem efficiently.
- The time complexity is optimal, and the code is well-structured and readable.
- The comments and variable names are clear and helpful.

Areas for Improvement:
- The space complexity can be optimized further by using variables to store only the previous row's values instead of maintaining the entire DP array. This would reduce the space complexity to O(1).
- The comment "Dp[i] the number of ways to make amout i" seems to be a copy-paste error from another problem and should be removed or corrected to reflect the current problem's context.

PR #1793, opened 2026-02-22
Add implementation for leetcode problems 256, 518

rishigoswamy:


super30admin:
Your solution for the Paint House problem is excellent. It correctly uses dynamic programming to achieve optimal time (O(n)) and space (O(1)) complexity. The code is clean and well-commented. 

One small suggestion: you don't need to compute `val` in every iteration. Instead, you can update `currCost` for each color and then at the end, return the minimum of `currCost`. This would make the code slightly more efficient by avoiding an unnecessary min operation in each loop. For example:

```
def minCost(self, costs: List[List[int]]) -> int:
    if not costs:
        return 0
    curr = costs[0]
        new_curr = [
            costs[i][0] + min(curr[1], curr[2]),
            costs[i][1] + min(curr[0], curr[2]),
            costs[i][2] + min(curr[0], curr[1])
        ]
        curr = new_curr
    return min(curr)
```

This change eliminates the `val` variable and only computes the min once at the end.

Note: You also included a solution for Coin Change II (leetcode_518.py) in your submission. While that solution is correct for its problem, it is not relevant to the Paint House problem. In the future, please ensure you only submit the solution for the problem being evaluated.

Overall, great job on the Paint House solution!

